PR #2, opened 2016-01-22
Allow to name files with dashes for the Ruby harness

eregon:
* SomeBenchmark -> some-benchmark.rb


smarr:
Toooo fancy for my taste, but well...

PR #30, opened 2016-12-25
Port benchmarks to Lua

fperrad:
Port all benchmarks to [Lua](http://www.lua.org/).

All the runtime stuff is available as Ubuntu/Debian packages:
- lua5.1	: Lua 5.1.x
- lua5.2	: Lua 5.2.x
- lua5.3	: Lua 5.3.x
- luajit	: LuaJIT 2.0.x which is ABI compatible with Lua 5.1 (they share same extensions)
- lua-bitop	: bitwise extension for Lua 5.1
- lua-socket	: micro-seconds resolution in benchmarks
- lua-check	: linter

[LuaJIT](http://luajit.org/) is a JIT trace compiler.
Before JIT compiling, LuaJIT acts as an interpreter which is 3x faster than a PUC Lua interpreter.
Without a large number of iterations, there is no JIT effect.


smarr:
Wow, this is really great. It will take me some time to go through all of it, but I'll try to comment as soon as possible.

Out of curiosity, did you run the number for instance for Java, and the different Lua implementations?

fperrad:
I tested them on Ubuntu 16.04 LTS with Lua 5.1.5, Lua 5.2.4, Lua 5.3.1 and LuaJIT 2.0.4

smarr:
I was more wondering about the performance numbers. Did you look into that, too? Just wondering.

fperrad:
No, I didn't that.

benchmarks/Lua/bounce.lua
```diff
@@ -0,0 +1,97 @@
...
+-- IMPLIED, INCLUDING BUT NOT LIMITED TO THE WARRANTIES OF MERCHANTABILITY,
+-- FITNESS FOR A PARTICULAR PURPOSE AND NONINFRINGEMENT. IN NO EVENT SHALL THE
+-- AUTHORS OR COPYRIGHT HOLDERS BE LIABLE FOR ANY CLAIM, DAMAGES OR OTHER
+-- LIABILITY, WHETHER IN AN ACTION OF CONTRACT, TORT OR OTHERWISE, ARISING FROM,
+-- OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN
+-- THE SOFTWARE.
+
+local Random = require'som'.Random
+
+local Ball = {_CLASS = 'Ball'} do
```
smarr: This `_CLASS = 'Ball'` isn't strictly necessary, or is it?
It seems to be used only in a few places, but not consistently. That's more for 'documentation', I presume?

benchmarks/Lua/bounce.lua
```diff
@@ -0,0 +1,97 @@
...
+    local random     = Random.new()
+    local ball_count = 100
+    local bounces    = 0
+    local balls      = {}
+    for i = 1, ball_count do
+        balls[i] = Ball.new(random)
+    end
+    for _ = 1, 50 do
+        for i = 1, #balls do
+            local ball = balls[i]
```
smarr: Hm, wouldn't this be better something like:

`for _, ball in pairs(balls) do`

benchmarks/Lua/towers.lua
```diff
@@ -0,0 +1,83 @@
...
+-- THE SOFTWARE.
+
+local towers = {} do
+setmetatable(towers, {__index = require'benchmark'})
+
+local function create_disk (size)
+    return {size = size, next = nil}
+end
+
+function towers:benchmark ()
```
smarr: I noticed that the whitespace is inconsistent from time to time.

Does a space belong here after `benchmark` or not?
There isn't one in a few other places, for instance on line 51 for `pop_disk_from`: https://github.com/smarr/are-we-fast-yet/pull/30/commits/fc2922c313a5a5c1f6f7ef6286093235dc1e70fd#diff-cb8e2d36e399aef9ea828138a6378a00R51

benchmarks/Lua/storage.lua
```diff
@@ -0,0 +1,54 @@
...
+end
+
+function storage:verify_result (result)
+    return 5461 == result
+end
+
+function storage:build_tree_depth (depth, random)
+    self.count = self.count + 1
+    if depth == 1 then
+        return {n = random:next() % 10 + 1}
```
smarr: If I am not mistaken, Lua doesn't have a table constructor that initializes the table for a specific size. So, this here is a good solution to ensue that the random function get's executed.

But, it is a little obscure. Would be great if a comment could be added with the reason.

benchmarks/Lua/permute.lua
```diff
@@ -0,0 +1,56 @@
...
+        for i = n, 1, -1 do
+            self:swap(n, i)
+            self:permute(n1)
+            self:swap(n, i)
+        end
+    end
+end
+
+function permute:swap (i, j)
+    self.v[i], self.v[j] = self.v[j], self.v[i]
```
smarr: This is fancy, and might be idiomatic, but, I'd prefer to keep the code as identical as possible:

```Lua
local tmp = self.v[i]
self.v[i] = self.v[j]
self.v[j] = tmp
```

benchmarks/Lua/nbody.lua
```diff
@@ -0,0 +1,180 @@
...
+    return bodies
+end
+
+function NBodySystem.new ()
+    local obj = {bodies = create_bodies()}
+    return setmetatable(obj, {__index = NBodySystem})
+end
+
+function NBodySystem:advance (dt)
+    for i = 1, #self.bodies do
```
smarr: Wouldn't here something like `for i, i_body in pairs(self.bodies) do` be nicer?

benchmarks/Lua/nbody.lua
```diff
@@ -0,0 +1,180 @@
...
+            i_body.vx = i_body.vx - dx * j_body.mass * mag
+            i_body.vy = i_body.vy - dy * j_body.mass * mag
+            i_body.vz = i_body.vz - dz * j_body.mass * mag
+
+            j_body.vx = j_body.vx + dx * i_body.mass * mag
+            j_body.vy = j_body.vy + dy * i_body.mass * mag
+            j_body.vz = j_body.vz + dz * i_body.mass * mag
+        end
+    end
+    for i = 1, #self.bodies do
```
smarr: Wouldn't here something like `for _, i_body in pairs(self.bodies) do` be nicer, too?

benchmarks/Lua/nbody.lua
```diff
@@ -0,0 +1,180 @@
...
+        local body = self.bodies[i]
+        body.x = body.x + dt * body.vx
+        body.y = body.y + dt * body.vy
+        body.z = body.z + dt * body.vz
+    end
+end
+
+function NBodySystem:energy ()
+    local e = 0.0
+    for i = 1, #self.bodies do
```
smarr: Wouldn't here something like `for i, i_body in pairs(self.bodies) do` be nicer, too?

benchmarks/Lua/bounce.lua
fperrad: I use 2 templates :
 - one for single object
```lua
local foo = {} do

function foo:bar ()
    -- do something
end

end -- object foo
```
- another for class
```lua
local Foo = {_CLASS = 'Foo'} do

function Foo.new ()
    local obj = {}
    return setmetatable(obj, {__index = Foo})
end

function Foo:bar ()
    -- do something
end

end -- class Foo
```
`_CLASS` is not mandatory, but useful when debugging because you could `print(obj._CLASS)`.
Using `_CLASS`  with a single object seems weird.
The `do` in first line and `end` in last line allow to create a scope for local variables.

benchmarks/Lua/towers.lua
fperrad: line 51 must be rewritten with whitespace
```lua
function towers:pop_disk_from (pile)
```

benchmarks/Lua/bounce.lua
fperrad: the function [`pairs`](https://www.lua.org/manual/5.3/manual.html#pdf-pairs) is an iterator for hash
```lua
for k, v in pairs(t) do
     -- do something
end
```

For sequence, (ie. an array starting at 1 and without hole) the best (fastest) way is
```lua
for i = 1, #t do
     local v = t[i]
     -- do something
end
```
The function [`ipairs`](https://www.lua.org/manual/5.3/manual.html#pdf-ipairs) is no longer recommended (lua-5.3.0-alpha was released without `ipairs`)
```lua
for i, v in ipairs(t) do
    -- do something
end
```

benchmarks/Lua/storage.lua
fperrad: Yes, a comment is needed

benchmarks/Lua/permute.lua
fperrad: Ok, I'll remove this idiom.

smarr:
I did run the numbers, and, well, I am little disappointed. I would have expected a bit better numbers for LuaJIT2:

fast-langs-overview-1

fperrad:
On http://awfy-speed.stefan-marr.de/comparison/ with only Lua53 and LuaJIT2,
I don't see the JIT effect (LuaJIT is only 3x faster than Lua53) with the benchmarks `DeltaBlue`, `Storage` and maybe `CD`.
Try a comparison between `luajit -joff` and `luajit`, and increase the number of inner iteration while the time is the same.
The options `-jv` and `-jdump` allow to see the behaviour of LuaJIT.

smarr:
For reference, the relevant chart: http://awfy-speed.stefan-marr.de/comparison/?exe=26%2BL%2Bmaster%2C25%2BL%2Bmaster&ben=22%2C21%2C9%2C14%2C13%2C20%2C10%2C11%2C18%2C15%2C17%2C16%2C12%2C19&env=1&hor=true&bas=25%2BL%2Bmaster&chart=normal+bars

And yes, these benchmarks seem to have issues. Haven't further looked into them yet.
`Storage` isn't really surprising: http://awfy-speed.stefan-marr.de/timeline/?exe=12%2C19%2C22%2C26%2C25%2C1%2C21%2C3&base=none&ben=%5Bpeak%5D+Storage&env=1&revs=50&equid=on

There were always plans to [improve the GC](http://wiki.luajit.org/New-Garbage-Collector), and Storage is a GC benchmark.

fperrad:
I added some minor improvements in another branch, see https://github.com/fperrad/are-we-fast-yet/commits/lua_improvement

fperrad:
Could you add more Lua variants ?
I am interested by comparison between :
1. LuaJIT 2.0.x and LuaJIT 2.1.x
2. Lua 5.1.x, Lua 5.2.x and Lua 5.3.x

benchmarks/Lua/som.lua
```diff
@@ -0,0 +1,631 @@
...
+function Set:collect (fn)
+    local coll = Vector.new()
+    for e in self:each() do
+        coll:append(fn(e))
+    end
+    return coll
+end
+
+function Set:contains (obj)
+    return self:has_some(function (it) return it == obj end)
```
smarr: Lua has only a single notion of equality? I suppose `==` is reference equality. There is no value/structural equality?

benchmarks/Lua/som.lua
```diff
@@ -0,0 +1,631 @@
...
+    end
+]])()
+end
+
+function Dictionary:is_empty ()
+    return self.size == 0
+end
+
+function Dictionary:get_bucket_idx (hash)
+    return (hash % self.buckets.n) + 1
```
smarr: using `%` here seems to be a performance drawback for Lua. I suppose the reason is to avoid another instance of code that's requiring code loading?

benchmarks/Lua/som.lua
```diff
@@ -0,0 +1,631 @@
...
+end -- class Entry
+
+local Dictionary = {_CLASS = 'Dictionary'} do
+
+local INITIAL_CAPACITY = 16
+local floor = math.floor
+
+function Dictionary.new (size)
+    local obj = {
+        buckets = {n = size or INITIAL_CAPACITY},
```
smarr: for the dictionary storage/buckets array, we also should try to make sure that it is a continuous array. Otherwise, performance for DeltaBlue and Havlak, I think, is likely very bad.

benchmarks/Lua/som.lua
```diff
@@ -0,0 +1,631 @@
...
+    end
+end
+
+function Dictionary:split_bucket (old_storage, i, head)
+    local lo_head, lo_tail = nil, nil
+    local hi_head, hi_tail = nil, nil
+    local current = head
+
+    while current do
+        if (floor(current.hash / old_storage.n) % 2) == 0 then
```
smarr: `floor() % 2` is likely also a performance issue.
Perhaps a solution would be to use a helper method `bitAnd` that's specialized based on the Lua version.
That's only a small overhead in the interpreter, and for the compiler it should be easy to eliminate.

benchmarks/Lua/som.lua
```diff
@@ -0,0 +1,631 @@
...
+    return setmetatable(obj, {__index = Dictionary})
+end
+
+function IdentityDictionary:new_entry (key, value, hash)
+    return IdEntry.new(hash, key, value, nil)
+end
+
+end -- class IdentityDictionary
+
+local LuaDictionary = {_CLASS = 'LuaDictionary'} do
```
smarr: This does not seem to be used?
I suppose you put it in for testing?

benchmarks/Lua/som.lua
```diff
@@ -0,0 +1,631 @@
...
+
+local Random = {_CLASS = 'Random'} do
+
+function Random.new ()
+    local obj = {seed = 74755}
+    return setmetatable(obj, {__index = Random})
+end
+
+function Random:next ()
+  self.seed = ((self.seed * 1309) + 13849) % 65536;
```
smarr: again, `bitAnd` performs likely better than `%`

smarr:
I am planning to add LuaJIT 2.1.x at some point, not sure about the other Lua variants, but shouldn't be a huge problem.

benchmarks/Lua/som.lua
fperrad: removed by https://github.com/smarr/are-we-fast-yet/pull/30/commits/92420f47606bd3c160bece0f7ca0723a0b46508d

benchmarks/Lua/som.lua
```diff
@@ -0,0 +1,631 @@
...
+    self.last_idx = self.last_idx + 1
+end
+
+function Vector:is_empty ()
+    return self.last_idx == self.first_idx
+end
+
+function Vector:each ()
+    local i = self.first_idx
+    local function iter ()
```
fperrad: done, see https://github.com/smarr/are-we-fast-yet/pull/30/commits/b3fa8cce1f1c335677ce1b2c22781994c880b8d6

benchmarks/Lua/som.lua
fperrad: refactored, see https://github.com/smarr/are-we-fast-yet/pull/30/commits/a981c7f062ecdbac641824cc9f421341a9b808f7

benchmarks/Lua/som.lua
fperrad: refactored, see https://github.com/smarr/are-we-fast-yet/pull/30/commits/a981c7f062ecdbac641824cc9f421341a9b808f7

benchmarks/Lua/som.lua
fperrad: refactored, see https://github.com/smarr/are-we-fast-yet/pull/30/commits/a981c7f062ecdbac641824cc9f421341a9b808f7

benchmarks/Lua/som.lua
```diff
@@ -0,0 +1,631 @@
...
+-- OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN
+-- THE SOFTWARE.
+
+local Vector = {_CLASS = 'Vector'} do
+
+local floor = math.floor
+
+function Vector.new (size)
+    local obj = {
+        storage   = {n = size or 50},
```
fperrad: see refactor in https://github.com/smarr/are-we-fast-yet/pull/30/commits/ac1eb74a63d9a29450e6c77fd78d3bfed4a835e4

benchmarks/Lua/som.lua
```diff
@@ -0,0 +1,631 @@
...
+    self.storage[idx] = val
+    if self.last_idx < idx + 1 then
+        self.last_idx = idx + 1
+    end
+end
+
+function Vector:append (elem)
+    if self.last_idx > self.storage.n then
+        -- Need to expand capacity first
+        self.storage.n = 2 * self.storage.n
```
fperrad: see refactor in https://github.com/smarr/are-we-fast-yet/pull/30/commits/ac1eb74a63d9a29450e6c77fd78d3bfed4a835e4

benchmarks/Lua/som.lua
```diff
@@ -0,0 +1,631 @@
...
+    end
+    return self.storage[idx]
+end
+
+function Vector:at_put (idx, val)
+    if idx > self.storage.n then
+        local new_n = self.storage.n
+        while idx > new_n do
+            new_n = new_n * 2
+        end
```
fperrad: see refactor in https://github.com/smarr/are-we-fast-yet/pull/30/commits/ac1eb74a63d9a29450e6c77fd78d3bfed4a835e4

benchmarks/Lua/som.lua
fperrad: see refactor in https://github.com/smarr/are-we-fast-yet/pull/30/commits/ac1eb74a63d9a29450e6c77fd78d3bfed4a835e4

benchmarks/Lua/som.lua
fperrad: Yes, only `==`.
The implementation of `IdentitySet` and `IdentityDictionary` are useless.
I could replace them by a final 
```lua
return {
    Vector = Vector,
    Set = Set,
    IdentitySet = Set,
    Dictionary = Dictionary,
    IdentityDictionary = Dictionary,
    Random = Random,
}
```
That just saves some lines of code.

fperrad:
@smarr ping

smarr:
@fperrad sorry, yeah, I should really get back to this. Will try to make time soon

smarr:
Just as an update, I set up Travis testing: https://github.com/smarr/are-we-fast-yet/tree/wip/lua
I might push that to this PR if that's ok.
Took longer than expected, so, I didn't make any progress on the actual review.

fperrad:
With Travis CI, a great tool is [hererocks](https://github.com/mpeterv/hererocks) which build/install Lua and Luarocks.

As example see https://github.com/fperrad/lua-TestLongString/blob/master/.travis.yml

smarr:
Sorry for taking soo much time.
Your PR is now finally merged.

Thanks a lot for the port!

PR #32, opened 2017-02-02
Add support for Squeak and RSqueak

timfel:


benchmarks/Smalltalk/Richards/Scheduler.som
```diff
@@ -31,18 +31,18 @@ Scheduler = RBObject (
       work: work
       state: state
       function: 
-        [:work :word |
-        | data functionWork | 
-        data := word.
-        functionWork := work.
+        [:thework :word |
```
smarr: Shadowing isn't supported?

benchmarks/Smalltalk/build-image-squeak.st
```diff
@@ -0,0 +1,65 @@
...
+
+Time class methodDict at: #primUTCMicrosecondsClock put: (Time class >> #utcMicrosecondClock).
+
+SmallInteger methodDict at: #'&' put: (SmallInteger >> #bitAnd:).
+SmallInteger methodDict at: #'%' put: (SmallInteger >> #\\).
+
+Smalltalk class compile: 'exit: code
+    ^ Smalltalk quitPrimitive' classified: '*SomLoader'.
+
+(Smalltalk classNamed: 'SomLoaderTests') compile: 'testLoadCDRedBlackTree
```
smarr: What's this?

benchmarks/Smalltalk/build-image-squeak.st
```diff
@@ -0,0 +1,65 @@
+| parentDir runner loader starting |
```
smarr: Could you please provide an explanatory comment on the PR about why this is needed, what you do here, and what changes are there that go beyond avoiding incompatibilities between Squeak/Pharo?

benchmarks/Smalltalk/run.st
```diff
@@ -1,5 +1,5 @@
 | args |
 "Adapt array to expectation of harness"
-args := Array with: nil withAll: Smalltalk arguments.
+args := {nil}, Smalltalk arguments.
```
smarr: Why this change? Is this tested with Pharo?

benchmarks/Smalltalk/Richards/Scheduler.som
timfel: No, the Squeak Compiler complains

benchmarks/Smalltalk/build-image-squeak.st
timfel: The entire what you might call scripting API and the new Filesystem classes do not exist in Squeak. Also, the default directory isn't the working dir, but the folder that contains the image. So the core of these changes just deal with finding out where the everything is and loading it.

benchmarks/Smalltalk/build-image-squeak.st
timfel: I found a bunch of code in the benchmarks and the harness that uses methods that are not available or work differently on Squeak. These lines above just patch the Squeak image to work with the loader (and work around a bug in the Windows file system classes)

timfel: Oh, and the test patches just insert "FileSystem slash" so the code works on Windows. But that method doesn't exist in Pharo anymore, AFAICT, so I'm just making the change here so that the test runs on Squeak. Maybe not the best place.

benchmarks/Smalltalk/run.st
timfel: with:withAll: doesn't exist in Squeak, but the literal array syntax also works in Pharo

benchmarks/Smalltalk/build-image-squeak.st
smarr: Hm, the `FileSystem slash` stuff, could that perhaps better go into the original code instead of doing it after the fact?

timfel: You're right, i will move it tomorrow

smarr: great, thanks!

smarr:
Thanks for looking at it, I'll not be able to get to merging it anytime soon though. Perhaps on the coming weekend. Sorry.

And, I'd still like to know whether on Squeak all the benchmarks run. Pharo/CogVM got issues with CD, if I remember correctly.

timfel:
Squeak runs CD, although it takes a very (very) long time. RSqueak currently hangs in CD, I was expecting rebench to kill it at some point, but that doesn't seem to be the case?

smarr:
Hm, ReBench kills stuff normally, it also prints a keep alive message.
Perhaps the timeout in the ReBench config is too long.

implementations/build-squeak.sh
```diff
@@ -0,0 +1,38 @@
...
+  cp Squeak5.1*.app/Contents/Resources/Squeak5.1*.image ../benchmarks/Smalltalk/Squeak.image
+  cp Squeak5.1*.app/Contents/Resources/Squeak5.1*.changes ../benchmarks/Smalltalk/Squeak.changes
+  cp Squeak5.1*.app/Contents/Resources/Squeak*.sources ../benchmarks/Smalltalk/
+
+  # This can be dropped once https://github.com/travis-ci/travis-build/pull/879 is merged and in production
+  if [[ "$(uname -s)" = "Linux" ]]; then
+      wget -q https://raw.githubusercontent.com/hpi-swa/smalltalkCI/master/utils/set_rtprio_limit.c
+      gcc -o set_rtprio_limit set_rtprio_limit.c
+      chmod +x ./set_rtprio_limit
+      sudo ./set_rtprio_limit $$
```
smarr: @timfel what is this for exactly? Is the required on all Linux systems or just travis?
When I want to run the benchmarks on the awfy benchmark server, I don't have sudo rights. And, I'd rather not try to mess with that. For Pharo, I don't have something like this. Is this related to the heartbeat VM thread? Can't we use the same VM (or VM type) as for Pharo? Would also be better for comparability of results.

timfel: Yes, this is related to heartbeat. I'll ping @fniephaus,but I remember they had long discussions about non-heartbeat VMs with Squeak, and there were some issues with it.

smarr: ok, so, to make the question more precise: Is there a difference between a heartbeat VM and one without when it comes to assessing the effectiveness of the CogVM JIT compiler?
Since that's the primary goal, everything that's merely related to UI etc should be ignored for the specific purpose of AWFY.

fniephaus: The high priority heartbeat thread is used to better capture input events AFAIK. So I'd say you could just use an itimer VM instead and remove that piece of code. Shouldn't affect the effectiveness of the JIT at all. Maybe even use [one of these](https://github.com/OpenSmalltalk/opensmalltalk-vm/releases/tag/201701281910).

smarr: Ok, I suppose that would be cog_linux32x86_squeak.cog.spur_201701281910_itimer.tar.gz then?

fniephaus:
Yes, sounds good. Let me know if there is any problem with this VM.

implementations/build-squeak.sh
fniephaus: Yes, sounds good. Let me know if there is any problem with this VM.

smarr:
@timfel I just noticed that there is a failing test when building the benchmark image, I think. Seems to be there form the beginning. And doesn't fail the build. So, I guess, it's not critical: https://travis-ci.org/smarr/are-we-fast-yet/jobs/207717183#L290

smarr:
@fniephaus looks good. Squeak works now on the benchmark server. RSqueak still needs a bit more work.

@timfel I would like to convert the RSqueak setup also into a source dependency that's properly versioned. At the moment, it seems you added just an unversioned bundle:
https://github.com/smarr/are-we-fast-yet/pull/32/files#diff-edf1b49f23e688c2baed5b6fcc377af7R15

I suppose to build RSqueak, it should be something like: `pypy .build/build.py --batch --32bit ` when all dependencies are available?

fniephaus:
On the other hand, you could also just use one of [these binaries](https://www.hpi.uni-potsdam.de/hirschfeld/artefacts/rsqueak/), the git commit info is available via --git-version.

smarr:
Tim set up the benchmarks based on these binaries, but I don't like it, because they aren't versioned, in the sense that they likely change when ever you decide to do a new build. So, I don't have a reproducible setup, and can decide myself when to update RSqueak.

Thanks for the tips, will look into it.

fniephaus:
No problem! Just dumping all that info on you, because I'm currently on mobile.
You could also use the binaries from the All-in-One bundle, because those VMs have passed our tests, and then update them whenever you wish.
Anyhow, I will of course leave all of this up to you :)

timfel:
@smarr I'd definitely recommend building a 64bit VM. It's what we use as our development focus, and the wordsize of the VM is independent from the image format

smarr:
Got the build working: https://github.com/smarr/are-we-fast-yet/blob/master/implementations/build-rsqueak.sh#L11

Just for documentation, nice to have for full reproducibility would be:

 - PyPy release with `not_rpython` and perhaps other things that 5.6.0 doesn't include it yet
 - RSqueak image ZIP (ideally versioned releases parallel to RSqueak

Just nice to have, at the moment, I don't think its necessary. And, I think you're doing your own performance tracking anyway.

smarr:
The results are in: http://awfy-speed.stefan-marr.de/timeline/#/?exe=19,21,13,29,28&base=none&ben=grid&env=1&revs=50&equid=off

I am not sure I made some mistake. The only benchmark that matches my expectations is Mandelbrot. Think the rest is more heavy on object/array access, and is pretty bad compared to Cog. Is that expected?

timfel:
Yes, object access is one of those things where we get massively better with maps and inlined fields and which is expected to be slower in many cases because there is extra indirection through storage strategies and for most objects integers will be stored wrapped. As I mentioned in the issue on the RSqueak repo, however, I am not inclined to merge the maps branch right now, because it makes BitBlt slower, and anything that makes the UI slower is a no-go for us at this time.

smarr:
Ok, yeah, was just surprised to see the results.

PR #33, opened 2017-02-02
Add Topaz Ruby

timfel:
Adds Topaz to the Ruby lineup. Not all benchmarks work, I think, but I'll get to that.

### DIFF
--- a/implementations/build-topaz.sh
+++ b/implementations/build-topaz.sh
@@ -5,9 +5,8 @@ source $SCRIPT_PATH/script.inc
 
 INFO Build Topaz
 pushd $SCRIPT_PATH
-  git clone git://github.com/topazproject/topaz
-  pushd topaz
-    git checkout tim/features
+  load_submodule $SCRIPT_PATH/Topaz
+  pushd Topaz
     pip install --upgrade invoke
     export TEST_TYPE=translate-jit-notest
     invoke travis.install_requirements
--- a/implementations/build-topaz.sh
+++ b/implementations/build-topaz.sh
@@ -1,0 +1,17 @@
+#!/bin/bash
+set -e # make script fail on first error
+SCRIPT_PATH=`dirname $0`
+source $SCRIPT_PATH/script.inc
+
+INFO Build Topaz
+pushd $SCRIPT_PATH
+  git clone git://github.com/topazproject/topaz
+  pushd topaz
+    git checkout tim/features
+    pip install invoke
+    export TEST_TYPE=translate-jit-notest
+    invoke travis.install_requirements
+    invoke travis.run_tests
+  popd
+popd
+OK Topaz Build Completed.
--- a/implementations/build-topaz.sh
+++ b/implementations/build-topaz.sh
@@ -8,7 +8,7 @@ pushd $SCRIPT_PATH
   git clone git://github.com/topazproject/topaz
   pushd topaz
     git checkout tim/features
-    pip install invoke
+    pip install --upgrade invoke
     export TEST_TYPE=translate-jit-notest
     invoke travis.install_requirements
     invoke travis.run_tests
--- a/rebench.conf
+++ b/rebench.conf
@@ -200,6 +200,9 @@ virtual_machines:
     RBX314:
         path: implementations/
         binary: rbx-314.sh
+    Topaz:
+        path: implementations/topaz/bin/
+        binary: topaz
 
     Crystal:
         path: implementations/
@@ -265,6 +268,8 @@ experiments:
             - JRubyJ8:
                 benchmark: ruby-indy
             - RBX314:
+                benchmark: ruby-rbx
+            - Topaz:
                 benchmark: ruby-rbx
     pharo:
         benchmark: pharo
--- a/rebench.conf
+++ b/rebench.conf
@@ -270,7 +270,7 @@ experiments:
             - RBX314:
                 benchmark: ruby-rbx
             - Topaz:
-                benchmark: ruby-rbx
+                benchmark: steady-ruby
     pharo:
         benchmark: pharo
         executions: Pharo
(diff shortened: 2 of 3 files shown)


rebench.conf
```diff
@@ -261,6 +264,8 @@ experiments:
                 benchmark: ruby-indy
             - RBX314:
                 benchmark: ruby-rbx
+            - Topaz:
+                benchmark: ruby-rbx
```
eregon: Is the `harness-rbx.rb` working better? Otherwise I'd recommend using `ruby-indy` or maybe `steady-ruby`.

timfel:
harness-rbx uses Time.now, the other uses the monotonic clock. We haven't implemented the latter for Topaz

timfel:
But yes, steady-ruby would otherwise seam like a better choice

implementations/build-topaz.sh
```diff
@@ -0,0 +1,27 @@
...
+    invoke travis.install_requirements
+    invoke travis.run_tests
+  popd
+
+  cat <<EOF>topaz.sh
+#!/bin/bash
+unset CDPATH
+pushd \`dirname \$0\`
+BINPATH="\$(pwd)"
+popd
```
eregon: How about simply having a checked-in file `implementations/topaz.sh` like e.g. [jruby.sh](https://github.com/smarr/are-we-fast-yet/blob/master/implementations/jruby.sh) ?

eregon:
Looks great, it would be awesome to have Topaz numbers for AWFY :smiley:

eregon:
> harness-rbx uses Time.now, the other uses the monotonic clock. We haven't implemented the latter for Topaz

I see, the monotonic clock would be quite useful as it's becoming standard in Ruby benchmarking tools.

implementations/build-topaz.sh
```diff
@@ -0,0 +1,17 @@
...
+set -e # make script fail on first error
+SCRIPT_PATH=`dirname $0`
+source $SCRIPT_PATH/script.inc
+
+INFO Build Topaz
+pushd $SCRIPT_PATH
+  git clone git://github.com/topazproject/topaz
+  pushd topaz
+    git checkout tim/features
+    pip install invoke
```
eregon: I needed to update my `invoke` with `pip install --upgrade invoke`, otherwise I got:
```
'travis.install_requirements' did not receive all required positional arguments!
```
Does it make sense to add the `--upgrade` flag here?

implementations/build-topaz.sh
```diff
@@ -0,0 +1,17 @@
...
+    export TEST_TYPE=translate-jit-notest
+    invoke travis.install_requirements
+    invoke travis.run_tests
```
eregon: This failed for me, I am not sure why:
```
[Timer] Timings:
[Timer] annotate                       ---   52.2 s
[Timer] rtype_lltype                   ---   98.6 s
[Timer] pyjitpl_lltype                 ---  412.7 s
[Timer] backendopt_lltype              ---   98.0 s
[Timer] stackcheckinsertion_lltype     ---   49.9 s
[Timer] database_c                     ---  138.1 s
[Timer] source_c                       ---   82.5 s
[Timer] compile_c                      ---   87.7 s
[Timer] ===========================================
[Timer] Total:                         --- 1019.7 s
./implementations/build-topaz.sh: line 15: run_tests: command not found
zsh: exit 127   ./implementations/build-topaz.sh
```

eregon:
I modified `test.conf` and it seems all benchmarks run on Topaz :smiley: , except Havlak which throws a `SystemStackError` in `do_dfs` (with a stack of 1787 frames, 356 recursive invocations of do_dfs):
```
	from /home/eregon/code/are-we-fast-yet/implementations/topaz/lib-topaz/range.rb:24:in `each'
	from /home/eregon/code/are-we-fast-yet/benchmarks/Ruby/som.rb:91:in `each'
	from /home/eregon/code/are-we-fast-yet/benchmarks/Ruby/havlak.rb:352:in `do_dfs'
	from /home/eregon/code/are-we-fast-yet/benchmarks/Ruby/havlak.rb:353:in `block in do_dfs'
	from /home/eregon/code/are-we-fast-yet/benchmarks/Ruby/som.rb:92:in `block in each'
	from /home/eregon/code/are-we-fast-yet/implementations/topaz/lib-topaz/range.rb:24:in `each'
```
Is there a way to augment the stack with Topaz?

implementations/build-topaz.sh
timfel: Yes, I'll add that

rebench.conf
timfel: I added the clock_gettime api and monotonic clock, now the default harness worky

timfel:
For small sizes all benchmarks should now run (had to fix a parser bug for json), but for adjusting the stacksize we don't have anything, yet.

timfel:
I've increased the default maximum stack size from 786kb to 2.8mb. Not sure if we also should do `ulimit -s unlimited` or something.

implementations/build-topaz.sh
timfel: Really not sure what the error is - translation has finished, I don't get why it is trying to run `run_tests` as command...

eregon: I think I modified the `build-topaz.sh` while it was building (to add the `--upgrade` flag), maybe `bash` has some strange reloading behavior. Anyway, let's ignore it for now.

smarr:
Thanks for looking at it, I'll not be able to get to merging it anytime soon though. Perhaps on the coming weekend. Sorry.

smarr:
Thanks! Merged it with a minor change to use a submodule for Topaz. That way, we can have deterministic runs. At least to a certain degree.

smarr:
I scheduled a run of MRI and Topaz on the benchmark server: https://github.com/smarr/awfy-runs/commit/80b12ed48b29e01354e32f64c47448b5ee2375e9

If everything works, I hope we'll have results later today.

PR #34, opened 2017-04-06
refactor Havlak

fperrad:
The repo https://github.com/hundt98847/multi-language-bench contains many language implementations of Havlak.

By comparison, I found over complicating in https://github.com/smarr/are-we-fast-yet.

I do the job only with Ruby, but I understand that these refactoring must be ported in all languages.

### DIFF
--- a/benchmarks/Lua/havlak.lua
+++ b/benchmarks/Lua/havlak.lua
@@ -266,10 +266,9 @@ function HavlakLoopFinder.new (cfg, lsg)
     local obj = {
         cfg = cfg,
         lsg = lsg,
-        non_back_preds = Vector.new(),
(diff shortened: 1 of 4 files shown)


smarr:
Hmm, I'll mark this as 'help wanted'. Generally, I don't think this is an absolutely necessary change. However, if someone wants to do it for all language, I am happy to accept it.

fperrad:
as https://github.com/smarr/are-we-fast-yet/pull/30 is merged, I add the refactor in Lua

smarr:
Hm, I think this isn't an essential improvement, and seems like there was not much interest in getting it consistent across languages. So, I am closing this. Though, if someone wants to take it one, I am happy to reconsider.

PR #35, opened 2017-07-03
alternate of som in Lua

fperrad:
Alternate implementation of som which does not follow the rules.


smarr:
Closing. Don't quite remember what the goal was here.

PR #36, opened 2017-07-04
[Lua] refactor with dofile

fperrad:
that avoids problem like previous commit (require + file naming)


smarr:
I don't remember what this was about. Sorry. Closing the PR, but can reconsider if this improves things.

PR #37, opened 2017-08-14
Add new result to CD

charig:
250 is very heavy for my benchmarking setting, 100 too fast for some implementations...


smarr:
Thanks, merged.

PR #41, opened 2018-10-14
Update CLBG link

fniephaus:
Same link as on corresponding Wikipedia page: https://en.wikipedia.org/wiki/The_Computer_Language_Benchmarks_Game


smarr:
Ok, so, I guess the original URL just went away, thanks for the fix.

fniephaus:
More info: https://www.reddit.com/r/programming/comments/88jg07/the_computer_language_benchmarks_games/

PR #45, opened 2019-05-03
Add TruffleSqueak (WIP)

fniephaus:
This PR adds GraalSqueak via GraalVM CE 1.0.0 RC16.

*WIP, do not merge...*

### DIFF
--- a/implementations/build-graalsqueak.sh
+++ b/implementations/build-graalsqueak.sh
@@ -1,0 +1,35 @@
+#!/bin/bash
+if [ "$1" = "style" ]
+then
+  exit 0
+fi
+
+set -e # make script fail on first error
+readonly SCRIPT_PATH="$(cd "$(dirname "${BASH_SOURCE[0]}")" && pwd)"
+source "${SCRIPT_PATH}/script.inc"
+
+readonly GRAALSQUEAK_DIR="${SCRIPT_PATH}/GraalSqueak"
+readonly GRAALSQUEAK_JAR="https://github.com/hpi-swa/graalsqueak/releases/download/awfy/graalsqueak-component.jar"
+readonly GRAALVM_VERSION=1.0.0-rc16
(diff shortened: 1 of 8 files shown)


implementations/build-graalsqueak.sh
```diff
@@ -0,0 +1,35 @@
...
+then
+  exit 0
+fi
+
+set -e # make script fail on first error
+readonly SCRIPT_PATH="$(cd "$(dirname "${BASH_SOURCE[0]}")" && pwd)"
+source "${SCRIPT_PATH}/script.inc"
+
+readonly GRAALSQUEAK_DIR="${SCRIPT_PATH}/GraalSqueak"
+readonly GRAALSQUEAK_JAR="https://github.com/hpi-swa/graalsqueak/releases/download/awfy/graalsqueak-component.jar"
```
smarr: Looks like this is not versioned. Think it might make it easier to have some version part of the string to be able to track changes.

fniephaus: Sure, that's one reason this PR is still WIP :)

smarr:
I merged #46, so, this one might need extra changes now?

fniephaus:
Yes, that's right...will update this soon.

smarr:
No need to hurry on my behalf 😄 I am not likely to get time to look at things in the next week.

smarr:
Hmmm. I suspect, this would need a new attempt.

I recently changed things so that there is no `implementations` folder anymore and everything is directly done in the language folder. In case of Squeak and Pharo, the `build.sh` does the major bits.

I also changed the CI setup to use `asdf`, since it seems like a good solution to also manage the language implementations on the benchmark servers.

So, if there's the interest, I'd suggest to check here:
- https://github.com/smarr/are-we-fast-yet/blob/master/.github/workflows/ci.yml#L60-L66
- https://github.com/smarr/are-we-fast-yet/blob/master/benchmarks/Smalltalk/build.sh#L9-L24

Thanks!

PR #52, opened 2020-10-30
optimized lua storage benchmark

ExtReMLapin:
before : 
```txt
Starting storage benchmark ...
storage: iterations=5000 average: 614us total: 3069000us

Total Runtime: 3069000us
```

after : 

```txt
Starting storage benchmark ...
storage: iterations=5000 average: 303us total: 1514000us

Total Runtime: 1514000us
```

ran on LuaJIT HEAD

### DIFF
--- a/benchmarks/Lua/storage.lua
+++ b/benchmarks/Lua/storage.lua
@@ -39,15 +39,15 @@ end
 function storage:build_tree_depth (depth, random)
     self.count = self.count + 1
     if depth == 1 then
-        -- With Lua, an array cannot be pre-allocated.
+        -- With Lua, an array cannot be pre-allocated. (actually yes, you can)
         -- We just compute the size like in others languages.
         return {n = random:next() % 10 + 1}
     else
-        local arr = {n = 4}
-        for i = 1, 4 do
-            arr[i] = self:build_tree_depth(depth - 1, random)
-        end
-        return arr
+        return {n = 4,
+        self:build_tree_depth(depth - 1, random),
+        self:build_tree_depth(depth - 1, random),
+        self:build_tree_depth(depth - 1, random),
+        self:build_tree_depth(depth - 1, random)}
     end
 end
 
--- a/benchmarks/Lua/storage.lua
+++ b/benchmarks/Lua/storage.lua
@@ -39,8 +39,7 @@ end
 function storage:build_tree_depth (depth, random)
     self.count = self.count + 1
     if depth == 1 then
-        -- With Lua, an array cannot be pre-allocated. (actually yes, you can)
-        -- We just compute the size like in others languages.
+        -- With Lua, an array can be preallocated with dummy values or with table.new in LuaJIT
         return {n = random:next() % 10 + 1}
     else
         return {n = 4,
--- a/benchmarks/Lua/storage.lua
+++ b/benchmarks/Lua/storage.lua
@@ -39,14 +39,14 @@ end
 function storage:build_tree_depth (depth, random)
     self.count = self.count + 1
     if depth == 1 then
-        -- With Lua, an array can be preallocated with dummy values or with table.new in LuaJIT
+        -- With Lua, an array can be preallocated with dummy values
         return {n = random:next() % 10 + 1}
     else
-        return {n = 4,
-        self:build_tree_depth(depth - 1, random),
-        self:build_tree_depth(depth - 1, random),
-        self:build_tree_depth(depth - 1, random),
-        self:build_tree_depth(depth - 1, random)}
+        local arr = {n = 4, true, true, true, true}
+        for i = 1, 4 do
+            arr[i] = self:build_tree_depth(depth - 1, random)
+        end
+        return arr
     end
 end
 


benchmarks/Lua/storage.lua
```diff
@@ -39,15 +39,15 @@ end
 function storage:build_tree_depth (depth, random)
     self.count = self.count + 1
     if depth == 1 then
-        -- With Lua, an array cannot be pre-allocated.
+        -- With Lua, an array cannot be pre-allocated. (actually yes, you can)
```
smarr, line 42: This doesn't seem to be a great comment any more, could you amend it perhaps to something more suitable?

benchmarks/Lua/storage.lua
```diff
@@ -39,15 +39,15 @@ end
...
         -- We just compute the size like in others languages.
         return {n = random:next() % 10 + 1}
     else
-        local arr = {n = 4}
-        for i = 1, 4 do
-            arr[i] = self:build_tree_depth(depth - 1, random)
-        end
-        return arr
+        return {n = 4,
+        self:build_tree_depth(depth - 1, random),
```
smarr, line 47: Would this still work if the array is allocated with something like `{n = 4, nil, nil, nil, nil}` to avoid the code duplication?

smarr (commented):
Thanks for the PR. I got two questions.

benchmarks/Lua/storage.lua
ExtReMLapin: Alright, removing this

benchmarks/Lua/storage.lua
ExtReMLapin: `{n = 4, nil, nil, nil, nil}`

"draws back" the performances from 

**~300ms to ~350ms**

While i'm about to say is purely subjective, Lua doesn't "actually" supports preallocation, you can fill it with nil values to prevent a ^2 reallocation when needed, when you don't know what the values will be.

My point is, fill it with `nil` is more like a hack than anything else.

I feel like using nils hack and later setting values inside with a loop is dirtier than the small code duplication.

GitSparTV: I'm testing `nil`s right now. I can say for sue it matters in PUC Lua, checking LuaJIT source code right now

GitSparTV: Ok, I don't know why does it speed up your benchmark, but according the LuaJIT `rehashtab` function the array is empty, so the size is shrinked to 0.
image
Here is `{nil,nil,nil,nil,nil,nil,nil,nil}`

image
here is `{false, false, false, false, false}`

I have a theory it changes the asize but not the allocated array itself

smarr:
Thanks for putting the effort into this.

And after sleeping over it, and looking through all the other implementations of the benchmark, I am slowly arriving at the view that this is likely not the right way to go. The real solution should be something along the lines of https://research.google/pubs/pub43823/

With respect to using FFI, I believe this is also not what these benchmarks should do. One can of course derive a variation that focuses on FFI performance. Though, this should then be very explicit about it.

To add to my explanation from earlier: different from other benchmark games, in the Are We Fast Yet benchmarks, I am not looking for the fastest way one could write these programs. Instead, I am looking to use a restricted subset of each language, that is as comparable as possible across languages, which in turn allows us to identify optimization potential in our runtime systems.

ExtReMLapin:
Smarr, i had a discussion with @GitSparTV, I told him I find the existence of this benchmark and the used implementation quite surprising.

Following this PDF : http://software.rochus-keller.ch/are-we-fast-yet_lua_results_2020-10-12.pdf

We find that most of the places where LuaJIT under-performs are the places where closures are used.

* Richards
* Havlak


`Storage` under-performed for no obvious reason, this is why I decided to fix the performances issues.


Comparing javascript, a technology built around the web and [asynchronous work/privacy] to Lua, a technology clearly not meant to be used in this context, seems wrong to me.

However, when we move to a pure-arithmetics benchmarks like mandelbrot, LuaJIT destroys V8 JIT compiler with no surprise.



>Instead, I am looking to use a restricted subset of each language, that is as comparable as possible across languages, which in turn allows us to identify optimization potential in our runtime systems.

But is it the right way ?


Some languages will have features that other will never have, for example, in Lua, you can fill the table on creature with `{n = 4, functioncall(), functioncall(), functioncall(), functioncall())`

Alright, you can't do it with javascript (or can you ?) but that's part of the languages war, there are some stuff that can and cannot be optimized automatically and the dev is supposed to make the correct steps to let the VM understand what is going or what is going to happen.

This is the point of static typing, as a developer, you do an extra step but it's for the greater good.


Shouldn't the implementation follow strictly pseudocode implementation like ? : 


`FILL AN ARRAY WITH 9 RANDOM INTEGERS`

`RANDOMIZE THE ARRAY ORDER`

Then do it in your language, the best way you can. I don't feel like benchmarking is just about the language, it's also about how the developer uses it.

Obviously, let's not fall into those *"beating C with 80 lines of haskell"* memes.

eregon:
As a 3rd point of view, I tend to agree with @smarr.
It seems the idiomatic way to create an array and fill it element by element in Lua is https://www.lua.org/pil/11.1.html,
assuming the length is not constant, or that duplicating the code for each element in the source is not wanted (this is the case here).
That said, I don't know much about Lua, so maybe I missed something?

Given this kind of code is likely to appear frequently in user code, it seems the job of the VM to optimize it.
For instance, if the example from the docs:
```lua
    a = {}    -- new array
    for i=1, 1000 do
      a[i] = 0
    end
```
is slow on LuaJIT, then that seems the real problem.

smarr:
My impression is we simply have different goals. I want to compare the runtime systems and compilers, not the languages.

The reasoning is detailed here: https://stefan-marr.de/papers/dls-marr-et-al-cross-language-compiler-benchmarking-are-we-fast-yet/

> Comparing javascript, a technology built around the web and [asynchronous work/privacy] to Lua, a technology clearly not meant to be used in this context, seems wrong to me.

Why would it be wrong? The languages benefit from very similar runtime systems and optimizations.
Again, for me, this is not about the languages. I am interested in implementation technology.

> But is it the right way ?

It is for the goal I have in mind, at least based on my current understanding (see for the above write up).

> you can fill the table on creature with

The syntax for instance for JavaScript is only minimally different, it works for objects and for arrays. Though, this is not the point. The point is that I want to stress specific parts of a runtime system, and ideally the same parts, as far as that is possibly between languages.

>  there are some stuff that can and cannot be optimized automatically

Doing research on this myself, the border is much further out than what one may think. There are a lot of know tricks that many implementations simply don't do. And yes, one reason is they often come with tradeoffs.

> Shouldn't the implementation follow strictly pseudocode implementation like ? :
> Then do it in your language, the best way you can. I don't feel like benchmarking is just about the language, it's also about how the developer uses it.

One can design benchmarks that way, and that's what the [Computer Language Benchmarks Game](https://benchmarksgame-team.pages.debian.net/benchmarksgame/index.html) does. Are We Fast Yet has very different goals. And I believe those to be useful, too.

krono:
I'd like to chime in.
In my view these benchmarks are not meant to answer the question “How fast can you possibly go?”.
But rather—with restrictions—“If performance was for free, how would you write it? And then, how fast is it?”.
I see it as an _idiomatic baseline_ and not merely C-in-disguise.

ExtReMLapin:
> For instance, if the example from the docs:
> 
> ```lua
>     a = {}    -- new array
>     for i=1, 1000 do
>       a[i] = 0
>     end
> ```
> 
> is slow on LuaJIT, then that seems the real problem.

Here i'm going to blame the developer, if you're writing this code when knowing you'll go with luajit, you're doing something wrong

smarr:
Thanks again for the contribution. At this point, I believe it's not something that is in alignment with the goals of my version of AWFY. As such, I'll close the PR.

That being said, I am sure there are other possible improvements, and I am happy to consider changes, as long as they fit within the project's vision.

Thank you

PR #56, opened 2021-05-14
Fix the return value of Queens#benchmark for the Ruby version

eregon:
* This seems to be the only Ruby benchmark with this issue.


smarr:
*sigh*

There's always something creeping past. Especially annoying since it's fixed in the Crystal version.

Thanks!

PR #57, opened 2021-05-16
Double literals and Comparisons in Ruby's CD

smarr:
This turns literals into doubles where it seems to match other benchmarks.
Turn the `==` in the red-black tree to `.equal?` since this seems to match better the semantics in the other benchmarks.

@eregon could you sanity check this?

### DIFF
--- a/benchmarks/Ruby/cd.rb
+++ b/benchmarks/Ruby/cd.rb
@@ -643,8 +643,8 @@ class CollisionDetector
     x = GOOD_VOXEL_SIZE * x_div
     y = GOOD_VOXEL_SIZE * y_div
 
-    x -= GOOD_VOXEL_SIZE if position.x < 0
-    y -= GOOD_VOXEL_SIZE if position.y < 0
+    x -= GOOD_VOXEL_SIZE if position.x < 0.0
+    y -= GOOD_VOXEL_SIZE if position.y < 0.0
 
     Vector2D.new(x, y)
   end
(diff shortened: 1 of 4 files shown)


benchmarks/Ruby/cd.rb
```diff
@@ -403,8 +403,8 @@ def right_rotate(y)
   end
 
   def remove_fixup(x, x_parent)
-    while x != @root && (!x || x.color == :black)
-      if x == x_parent.left
+    while !x.equals(@root) && (!x || x.color == :black)
```
eregon, line 406: `.equal?`

eregon:
> This turns literals into doubles where it seems to match other benchmarks.

Looks good.

> Turn the `==` in the red-black tree to `.equal?` since this seems to match better the semantics in the other benchmarks.

What's the rationale for this one? Semantically it's the same unless the nodes define `==` (they don't).
I think `equal?` is rarely used explicitly in Ruby, so `==` seems more idiomatic.

benchmarks/Ruby/cd.rb
eregon: Weird that the test run in CI doesn't catch that, is this never executed, or only with more iterations?

smarr: There's some code not reached in the rotations logic I think.
Though, I thought, I did a search and replaces for `.equals?` hmpf. I made that mistake in other places.

smarr:
> What's the rationale for this one? Semantically it's the same unless the nodes define == (they don't).
> I think equal? is rarely used explicitly in Ruby, so == seems more idiomatic.

There were some other `equal?` already, and the code looked inconsistent in the usage of `==` vs. `.equal?` https://github.com/smarr/are-we-fast-yet/blob/master/benchmarks/Ruby/cd.rb#L175

The Java version uses `==`, and the Python version now uses `is` to match that.
So, my intention was mostly to make the Ruby code consistent. Looking the the difference between `==`, `.eq?`, and `.equal?`, it seemed `.equal?` would be the best match for pointer equality.

I am find with using `==` if you think it's more idiomatic, and tell me that semantics-wise it's not an issue :)

eregon:
It's probably good to match other languages then and use `equal?` explicitly for this case.
I think in Ruby people would almost only use `equal?` if they know `==` is redefined (e.g. comparing by value).
It doesn't matter much either way, in fact both methods end up calling the target: `method(:equal?) == method(:==) # => true`.

PR #62, opened 2021-10-21
Remove unused temp

fniephaus:


smarr:
No idea why Squeak is failing. Looks like a network fluke.

PR #63, opened 2021-11-09
Add Julia as a language

vchuravy:


smarr:
Thanks @vchuravy, Julia would be a very interesting addition.

Please note the reasoning and rules behind these benchmarks:
 - https://stefan-marr.de/papers/dls-marr-et-al-cross-language-compiler-benchmarking-are-we-fast-yet/
 - https://github.com/smarr/are-we-fast-yet/blob/master/docs/guidelines.md
 - https://github.com/smarr/are-we-fast-yet/blob/master/docs/core-language.md

They may or may not fit your specific needs and interests, though, they are comparably strict to enable specific kinds of comparisons.

I haven't programmed in Julia before, and only know the language from academic papers on its implementation. Would be curious how you'll resolve some of the quirks in the benchmarks.

vchuravy:
I will try to stick to the structure, and yes the goal is to be comparative :)

PR #65, opened 2022-03-24
Add C# language

jfheins:
Hi,
I thought you might like to add C# to the mix. I used .net 6 so it should be cross platform and run under linux, mac and windows.

If you think C# is a valuable addition, I will continue to port the java benchmarks over. Maybe 2 questions on the code I already wrote:

- Does "System.Numerics.Vector<double>" count as a undesired data structure? (It does give a 20% speedup)
- I refactored the momentum calculation to `bodies.Aggregate(new Vector<double>(), (m, b) => m + b.Velocity * b.Mass);` which uses a lambda but is less identical to java. ok or revert?

Just trying to get a feeling for your goals before I tacke the other benchmarks.


smarr:
Hi @jfheins, yes, the C# benchmarks would be very valuable.

I do have some initial versions of the following already:
 - Bounce
 - List
 - Permute
 - Queens
 - Sieve
 - Storage
 - Towers

> Does "System.Numerics.Vector" count as a undesired data structure? (It does give a 20% speedup)

Yes, I would avoid it. The idea would be to use data structures that are common to "all" languages. So, we would want to restrict ourselves to plain arrays where possible.

For the NBody benchmark, it is also surprising to me that the compiler can't figure it out.

> I refactored the momentum calculation to bodies.Aggregate(new Vector<double>(), (m, b) => m + b.Velocity * b.Mass); which uses a lambda but is less identical to java. ok or revert?

When in doubt about these things, I'd suggest to look at the range of choices we make for other languages. Though, I'd probably want to stick with what Java does. I'd also would want to keep each of the coordinate dimensions as a separate property. Your choice to turn them into vectors prevents use to assess the getter/setter performance, which is one of the main bits measured by the NBody benchmark.

jfheins:
I ported Richards now as well, I think this could go in if style is fine. I saw the csharp branch, should I try to merge the two or would you rather do it yourself?

benchmarks/JavaScript/richards.js
```diff
@@ -22,7 +22,7 @@ var NO_TASK = null,
 
   DATA_SIZE = 4,
 
-  TRACING = false;
+  TRACING = true;
```
smarr, line 25: Please don't change this. It's not a bug.

benchmarks/JavaScript/richards.js
```diff
@@ -257,7 +257,7 @@ Scheduler.prototype.trace = function (id) {
     process.stdout.write("\n");
     this.layout = 50;
   }
-  process.stdout.write(id);
+  process.stdout.write(id.toString());
```
smarr, line 260: This is dead code. I'd not touch it, but am fine either way.

benchmarks/Csharp/Benchmarks/NBody.cs
```diff
@@ -0,0 +1,195 @@
+﻿namespace Harness.Benchmarks;
```
smarr, line 1: Hm, perhaps the namespace could be `AreWeFastYet` or so?

benchmarks/Csharp/.editorconfig
```diff
@@ -0,0 +1,223 @@
+# Remove the line below if you want to inherit .editorconfig settings from higher directories
```
smarr, line 1: Is this strictly needed?

benchmarks/Csharp/Benchmarks/NBody.cs
```diff
@@ -0,0 +1,195 @@
...
+
+public sealed class NBody : IBenchmark
```
smarr, line 3: Is the `I` here required by naming convention?
Benchmark shouldn't be an interface but a class. 
There's a concrete method in the `Benchmark` class.

benchmarks/Csharp/Benchmarks/NBody.cs
```diff
@@ -0,0 +1,195 @@
...
+{
+    public bool Benchmark(int innerIterations)
+    {
+        var system = new NBodySystem();
+        for (var i = 0; i < innerIterations; i++)
+        {
+            system.Advance(0.01);
+        }
+
+        return VerifyResult(system.Energy(), innerIterations);
```
smarr, line 13: This is not the same code as it's in any of the other versions.
Please stick closely to that code.

benchmarks/Csharp/Benchmarks/NBody.cs
```diff
@@ -0,0 +1,195 @@
...
+        }
+
+        foreach (var body in bodies)
+        {
+            body.X += dt * body.Vx;
+            body.Y += dt * body.Vy;
+            body.Z += dt * body.Vz;
+        }
+
+        void CalculateVelocities(Body a, Body b)
```
smarr, line 80: Is there a reason to not follow the code precisely? Any deviation will invalidate any comparison.

benchmarks/Csharp/Benchmarks/NBody.cs
```diff
@@ -0,0 +1,195 @@
...
+        X = x;
+        Y = y;
+        Z = z;
+        Vx = vx * DAYS_PER_YER;
+        Vy = vy * DAYS_PER_YER;
+        Vz = vz * DAYS_PER_YER;
+        Mass = mass * SOLAR_MASS;
+    }
+
+    public double X { get; set; }
```
smarr, line 144: It's not sufficient to define this as a public field?

benchmarks/Csharp/Benchmarks/Rich/IdleTaskDataRecord.cs
```diff
@@ -0,0 +1,14 @@
+﻿namespace Harness.Benchmarks.Rich;
```
smarr, line 1: Please don't abbreviate the benchmark name in the folder name.

It's also not clear that you follow the same code style across benchmarks. Wouldn't it be more consistent to either make all classes separate files, or have all benchmarks each in a single file?

benchmarks/Csharp/Benchmarks/Rich/Scheduler.cs
```diff
@@ -0,0 +1,290 @@
...
+        TaskControlBlock t = FindTask(identity);
+        if (NO_TASK == t)
+        {
+            return NO_TASK;
+        }
+        t.IsTaskHolding = false;
+        return t.Priority > currentTask.Priority ? t : currentTask;
+    }
+
+    [Conditional("TRACING")]
```
smarr, line 252: Please don't do anything fancy like this.
Please stick to the basic core language.

benchmarks/Csharp/Program.cs
```diff
@@ -0,0 +1,105 @@
...
+IBenchmark? CreateBenchmarkInstance(string name)
+{
+    var benchmarkClass = Type.GetType("Harness.Benchmarks." + name, false, true);
+    if (benchmarkClass == null)
+        return null;
+    var benchmarkInstance = Activator.CreateInstance(benchmarkClass) as IBenchmark;
+    return benchmarkInstance;
+}
+
+interface IBenchmark
```
smarr, line 50: Please stick with how `Benchmark` is implemented in other languages, or in the CSharp branch.

benchmarks/Csharp/Program.cs
```diff
@@ -0,0 +1,105 @@
...
+{
+    public bool Benchmark(int innerIterations);
+}
+
+class Run
+{
+    public int Iterations { get; set; } = 1;
+    public int InnerIterations { get; set; } = 1;
+    public TimeSpan TotalRuntime { get; private set; }
```
smarr, line 59: Please stick to the types used in other languages where possible.

smarr (commented):
A few comments. There are some things that do not seem to match the original implementation and may cause undesired aspects to be included in any comparison.

benchmarks/JavaScript/richards.js
jfheins: will revert

benchmarks/JavaScript/richards.js
jfheins: Sorry for the other change, this was the bug I wanted to fix. On my node the previous code gives me

> TypeError [ERR_INVALID_ARG_TYPE]: The "chunk" argument must be of type string or an instance of Buffer or Uint8Array. Received type number (123)

benchmarks/Csharp/.editorconfig
jfheins: It freezes linter settings but ... not stricly needed, no.

benchmarks/Csharp/Benchmarks/NBody.cs
jfheins: I is naming convetion for an Interface, yes. But I can make Benchmark an abstract class (ideally generic in return type to avoid the object casting) and then it will be `Benchmark`.

benchmarks/Csharp/Benchmarks/NBody.cs
jfheins: Look like my refactoring habit took over ;-)
Will change this

benchmarks/Csharp/Benchmarks/NBody.cs
jfheins: That would be perfectly sufficient, although it would also be against MS ""

Java also used the property pattern i.e.:
```
  private double x;
  public double getX() { return x; }
  public void setX(final double x) { this.x = x; }
```
C# property is the equivalent for Java field with get & set methods

benchmarks/Csharp/Benchmarks/Rich/IdleTaskDataRecord.cs
jfheins: I had that, but my problem was that C# cannot handle a class and a Sub-Namespace with the same name.
I already had a class "Richards" in the namespace `Harness.Benchmarks` so I renamed the sub namespace. (Note that namespaces need not correlate with folder structure in C#)

As Richards has much more code than nbody, I felt it might need more structure. If you think they should use the same style, which one would you prefer?

benchmarks/JavaScript/richards.js
smarr: Ok, thanks. Fixing it would be good then :)

benchmarks/Csharp/.editorconfig
smarr: Ah, ok, which linter are you using, and would it be possible to add it to the GitHub Actions setup?
If it's for a linter, having it would be good, yes.
I was just wondering what it was since the file name is `.editorconfig` instead of check style, or lintrc or similar.

Thanks!

benchmarks/Csharp/Benchmarks/NBody.cs
smarr: Hm, I'd be worried about the implications from a compilation perspective.

This seems to be a change from other implementations, and it's not clear to me that the departure is "necessary" in a sense.

AreWeFastYet only "works" because we are trying to be very consistent, and balance idiomatic constraints with comparability/fairness.
So, ideally, we give each language's compiler/runtime exactly the same challenges, minimizing any deviation that's not forced on us by language differences.

benchmarks/Csharp/Benchmarks/NBody.cs
smarr: I'll happily defer to your or the linter's judgement of what's idiomatic.

My thought was, especially when compared to Java: in C# you don't need to change the client code when switching from `public double X` to `public double X {...}`, so, a field might be fine. But indeed, hopefully the compiler will eliminate the difference for trivial ones already anyway. So, using a full property is also a good test to make sure the compiler actually does it.

benchmarks/Csharp/Benchmarks/Rich/IdleTaskDataRecord.cs
smarr: Yeah, it's more code, but if we can have a single file per benchmark that would match better what we have for other benchmarks. So, it's a bit of an attempt to get cross-language consistency. Helps with maintaining all the languages... especially when bugs are found, which gets tedious with the increasing number of languages :)

So, avoiding variation is a good thing to keep things going.

benchmarks/Csharp/.editorconfig
jfheins: I was using the Visual Studio 2022 integrated hints, like these:
image
For better linting (especially reformatting) we are using Resharper at my main job. The cli tool is free of charge and can be integrated into github actions. (and it can load the editorconfig settings)

smarr: Thanks, if Resharper is an option, would be great to have similar to CheckStyle, PyLint, RuboCop and the other linter setups :)

benchmarks/Csharp/Benchmarks/NBody.cs
jfheins: From a compilers perspective, I think this is a virtual call (callvirt IL instruction) one way or the other. But I see that you are favoring consistency stringly and will change this to be a 1:1 match with Java

benchmarks/Csharp/Benchmarks/NBody.cs
jfheins: Changing a field to a property in C# is not breaking if (and only if) you can re-compile all the consumers. (or it is source-compatible but binary-breaking)
As far as I know, the binary breaking change is the reason MS recommends properties.

benchmarks/Csharp/Program.cs
jfheins: fixed

benchmarks/Csharp/Program.cs
jfheins: fixed. I added a comment that it is microseconds though

benchmarks/Csharp/Benchmarks/NBody.cs
smarr: Ah, right, very good reason to go with properties then :)

benchmarks/Csharp/Program.cs
smarr: Great :) 👍🏻 
Thanks!

jfheins:
I merged your branch in to extend the test suite - I hope that's ok? (The brnach also contained some changes outside of the CSharp folder)

In the Harness, I added a Dictionary that keeps track of all known Benchmarks, this might be a tad longer but it is also less "magic" than using Reflection so I'd expect more maintainable for you. (Reflection was easy to break with the namespaces)

smarr:
I did rebase the PR and started setting up ReSharper as well as basic compiling and running on GitHub Actions. May try to look at the current benchmarks, some style issues.

huahaofeng:
@smarr Is this PR eligible to merge?

smarr:
> @smarr Is this PR eligible to merge?

Why do you ask? I had started the last missing benchmark, Havlak, a while ago, but its still not finished. So, the PR is not quite ready yet.

If you would like to contribute, you're very welcome.

If you have a specific use case, it would be great to know, too.

nikolahua:
> > @smarr Is this PR eligible to merge?
> 
> Why do you ask? I had started the last missing benchmark, Havlak, a while ago, but its still not finished. So, the PR is not quite ready yet.
> 
> If you would like to contribute, you're very welcome.
> 
> If you have a specific use case, it would be great to know, too.

I have been looking for benchmark suite which is used to measure performance between different versions of .NET runtime for a long time. I've got nothing until I find [this answer](https://www.quora.com/Is-the-Mono-CLR-really-slower-than-CoreCLR). However, the results given by this answer seem to be based on the programming language named Oberon. It will become a very valuable benchmark suite for .NET if the C# benchmarks are also added in.

jfheins:
Commit 29755eaeef2f8c2f982ba25bcd10afa4a330574d is one you may want to undo; your choice.

As far as know, earlier Java implemented substring as a second view into the same bytes so O(1) and that was changed at some point. C# always had O(n) substrings.
However, when Span<char> was added to C#, it basically gained a O(1) readonly substring. This is significantly quicker, but likely not the apples-to-apples comparison that you want to have.

71275e1e3bd5da639e60b60ae6e52099631ba07b is a relevant bugfix though as the parameters differ between Java and C#.

image

jfheins:
Also Havlak seems to benefit from Server GC in dotnet

benchmarks/CSharp/Json.cs
```diff
@@ -0,0 +1,1004 @@
...
+    {
+      return new ParseException(message, index, line, column - 1);
+    }
+
+    private bool IsWhiteSpace()
+    {
+      return current == ' ' || current == '\t' || current == '\n' || current == '\r';
+    }
+
+    private bool IsDigit()
```
jfheins, line 985: Just fyi there is already a [char.IsDigit()](https://learn.microsoft.com/en-us/dotnet/api/system.char.isdigit) method in .net

benchmarks/CSharp/Json.cs
```diff
@@ -0,0 +1,1004 @@
...
+
+      return Error("Expected " + expected);
+    }
+
+    private ParseException Error(string message)
+    {
+      return new ParseException(message, index, line, column - 1);
+    }
+
+    private bool IsWhiteSpace()
```
jfheins, line 980: char.IsWhiteSpace also exists.

The build-in methods are more unicode aware, so there is a semantic difference

PR #67, opened 2022-06-21
Ruby harness improvements

eregon:
This is needed to integrate with polybench, cc @boris-spas.

Notably this makes it possible to run a benchmark without needing to run harness.rb, like:
```ruby
require_relative 'run'

RUN = Run.new('Bounce')
RUN.num_iterations = 1
RUN.inner_iterations = 1

5.times do
  RUN.run_benchmark
end
```


eregon:
@smarr In CD, RuboCop is suggesting to change:
```ruby
    # Y is guaranteed to be non-null at this point.
    if y.left
      x = y.left
    else
      x = y.right
    end
```
to
```ruby
    # Y is guaranteed to be non-null at this point.
    x = y.left || y.right
```

I do that, or I add an ignore? (it is semantically equivalent, it avoids calling `y.left` twice)

smarr:
> ```ruby
>     # Y is guaranteed to be non-null at this point.
>     x = y.left || y.right
> ```
> 
> I do that, or I add an ignore? (it is semantically equivalent, it avoids calling `y.left` twice)

Hm, my feeling is that we would want a compiler to figure this out. So, I'd tent to disabling the relevant rubocop rule.

benchmarks/Ruby/.rubocop.yml
```diff
@@ -50,7 +50,6 @@ Layout/LineLength:
     - 'json.rb'
     - 'havlak.rb'
     - 'harness.rb'
-    - 'harness-rbx.rb'
```
smarr, line 53: Hm, this file should be deleted now?

smarr (approved):
Just for confirmation, this is just moving code around right?

@boris-spas mentioned in Potsdam that he indeed wanted to do that to replicate the same code structure as in the other languages.

So, yeah, that's great.

Thanks

eregon:
> Just for confirmation, this is just moving code around right?

Mostly, yes. If you review commit-by-commit it's clearer what happens and what are the actual changes.

eregon:
@boris-spas will double-check it works for polybench and then it should be ready to merged (that's why it's a draft PR currently).

benchmarks/Ruby/harness-rbx.rb
```diff
@@ -55,7 +55,7 @@ def load_benchmark_suite(benchmark_name)
       benchmark_file = benchmark_name.gsub(/([a-z])([A-Z])/) { "#{$1}-#{$2}" }.downcase
     end
     unless require_relative(benchmark_file)
-      raise "failed loading #{benchmark_file}"
```
smarr, line 58: Hm, is this really better? What happens if you name a benchmark that doesn't exist?

eregon: `require_relative` raises a LoadError in that case (as the commit message says)

eregon: Here are the docs just in case:
```
=== Implementation from Kernel
------------------------------------------------------------------------
  require_relative(string) -> true or false

------------------------------------------------------------------------

Ruby tries to load the library named string relative to the
requiring file's path.  If the file's path cannot be determined a
LoadError is raised. If a file is loaded true is returned and false
otherwise.
```
They are pretty bad :/ and I understand your confusion now.
`require` docs are a bit clearer:
```
=== Implementation from Kernel
------------------------------------------------------------------------
  require(name)    -> true or false

------------------------------------------------------------------------

Loads the given name, returning true if successful and false if the
feature is already loaded.
...
 If the file named cannot be found, a
LoadError will be raised.
```

eregon: https://github.com/ruby/ruby/commit/1f067e775d8629ac3b33f96dd885d5f996e787f3

boris-spas:
LGTM! Polybench happily works with
```ruby
require(PATH_TO_AWFY_BENCHMARKS_RUBY + "/run")

RUN = Run.new('Bounce')
RUN.num_iterations = 1
RUN.inner_iterations = 1

def run()
    RUN.run_benchmark
end
```

eregon:
Ready to be merged then

smarr:
Thank you!

PR #69, opened 2022-07-29
Fix CD benchmark for Smalltalk.

fniephaus:
The additional code in the Smalltalk version of `CollisionDetector>>isInVoxel:motion:` that forces IEEE floating point semantics used `Float negativeInfinity` instead of `Float infinity` for `low_x` and `low_y`. This has accidentally caused a max stack height of more than 500.000 (as opposed to 5), which in turn caused stack overflow errors in TruffleSqueak. With this fix in place, we can finally run the CD benchmark on TruffleSqueak.

### DIFF
--- a/benchmarks/Smalltalk/CD/CollisionDetector.som
+++ b/benchmarks/Smalltalk/CD/CollisionDetector.som
@@ -94,9 +94,9 @@ CollisionDetector = (
     y0  := init y.
     yv  := fin y - init y.
 
-    xv = 0.0
+    xv = 0.0 "follow IEEE floating point semantics"
       ifTrue:  [
-        low_x  := Float negativeInfinity.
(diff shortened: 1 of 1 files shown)


smarr (approved):
Thank you!

fniephaus:
According to the simple one-short runs on GitHub actions, the CD benchmarks now runs in ~24ms on Squeak, which is exactly 0.08% of the time (~29658ms) it ran before the fix was in place. :tada:

Before: https://github.com/smarr/are-we-fast-yet/runs/7329854689?check_suite_focus=true#step:17:125
After: https://github.com/smarr/are-we-fast-yet/runs/7577058390?check_suite_focus=true#step:17:131

smarr:
Yepp, saw that :) Thank you for debugging this!

Issue #72, opened 2022-08-16
Should frozen_string_literal be used for Ruby benchmarks?


eregon:
For instance on `Json` it seems like it would make sense given there are lots of String literals, which are actually allocations without `# frozen_string_literal: true` as first line.
From a quick local run, it's faster on CRuby with `frozen_string_literal` and it is about the same for TruffleRuby:
```
CRuby: ~428091 -> ~290305
TruffleRuby: 66057us -> 65282us
```

OTOH, maybe we want to measure how well the VM optimizes?

Ruby code in the wild is a mix of `# frozen_string_literal: true` and not using it. Ruby 3 still isn't using that by default over compatibility concerns. I'd think nowadays for performance-sensitive files people typically use `# frozen_string_literal: true`.

smarr:
Hmmm. So, I did actually add those to the copy I am using for tracking TruffleRuby performance.

https://github.com/smarr/truffleruby-faststart/commit/f2a232426ca6d399a4f7d61799fb893b5f89ed1b

Not sure why I didn't propagate it back here.

eregon:
Right, at least for interpreter performance I think adding it makes a lot of sense.
Could you apply it to this repo too?

PR #73, opened 2022-08-16
[Ruby] Apply `# frozen_string_literal: true` to all benchmarks

smarr:
This PR changes the Ruby benchmarks to use frozen, i.e., immutable string literals.

It seems good engineering practice to have immutable strings, and it may enable further optimizations.

Arguably, this may even be more comparable to other languages, since strings are often immutable.

On the other hand, one may want a VM to figure out automatically that things are immutable to benefit from it. So, there’s a tradeoff here.

For the moment, it seems useful to go with the immutable version, which may become the default in Ruby at some point, too.

This addresses #72.

@eregon does this look alright to you?

### DIFF
--- a/benchmarks/Ruby/benchmark.rb
+++ b/benchmarks/Ruby/benchmark.rb
@@ -1,3 +1,5 @@
+# frozen_string_literal: true
+
 # Copyright (c) 2015-2016 Stefan Marr <git@stefan-marr.de>
 #
 # Permission is hereby granted, free of charge, to any person obtaining a copy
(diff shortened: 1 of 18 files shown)


eregon (approved):

eregon (approved):

eregon:
Looks good.
I'm unsure if Crystal also has `frozen_string_literal`, if so seems best to add it there too.

smarr:
Crystal's strings seem to be immutable by default:

"A [String](https://crystal-lang.org/api/1.5.0/String.html) represents an immutable sequence of UTF-8 characters."

PR #77, opened 2022-10-12
Update JS to Modern ECMAScript, and add AOT support, also for Java

smarr:
This updates the JavaScript to use ECMAScript classes.

It also includes a WebPack configuration to generate a single JS file.
Just run `npx webpack` and use the `/dist/harness.js`.

@manuel-serrano this hopefully avoids various outdated JS idioms.


smarr:
Updated README a little bit.

PR #78, opened 2022-10-25
Add Babel to Webpack config to be able to compile to ECMAScript 5

smarr:
Just running `npm install . && npm run webpack` will generate a plain es5 compatible dist/harness.js file.

Without babel, we would get a single file version of AWFY, which may also be useful for some runtime systems.


smarr:
The last change adapts the webpack config to produce ES5 and ES2022 files, and adds a README.

PR #98, opened 2025-02-07
Fix off-by-one in Havlak result verification.

vext01:
Before, if you used a parameter with no "canned result", you would get:
```
$ lua53 harness.lua Havlak 1 2
Starting Havlak benchmark ...
No verification result for 2 found
lua53: ./havlak.lua:640: bad argument #1 to 'format' (number expected, got nil)
```


smarr (approved):
Good catch. Thank you!

PR #99, opened 2025-02-12
Use `ifNil:` in `isShorter:than:`.

fniephaus:
I'd argue this is slightly more idiomatic and it seems that this is also [supported in SOM](https://github.com/SOM-st/SOM/blob/f11313b14e6b85c1b5c5567e5b8cedc4b8e814fd/Smalltalk/Object.som#L51). In fact, the Squeak/Smalltalk compiler turns `ifNil:` into an identity check (`== nil`) plus jump rather than sending `isNil` to the receiver and then jump.

I noticed this some time ago when I looked into excessive splitting in TruffleSqueak ([here's 
a recent example](https://gist.github.com/TruffleSqueak-Bot/0fb5cbff18d9f47b3305a4d1879aa2bc#file-9ce8a19c-list-trace-log)). It turned out that the additional `isNil` send causes these splits.

Surely, there are ways to fix the splitting issue in TruffleSqueak. I just thought I open this anyway because, again, it is more idiomatic and because of that, also specifically optimized by the Squeak/Smalltalk compiler.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -13,7 +13,7 @@ jobs:
           - { name: SOM, id: som, folder: SOM }
           - { name: Crystal, id: crystal, folder: Crystal }
           - { name: JavaScript, id: js, folder: JavaScript }
-          - { name: SOMns, id: somns, folder: SOMns, ubuntu: ubuntu-20.04 }
+          - { name: SOMns, id: somns, folder: SOMns, ubuntu: ubuntu-24.04 }
           - { name: Pharo, id: pharo, folder: Smalltalk }
           - { name: Squeak, id: squeak, folder: Smalltalk }
           - { name: Ruby, id: ruby, folder: Ruby }
@@ -119,9 +119,11 @@ jobs:
       - name: Install SOMns
         run: |
           source benchmarks/script.inc
-          load_git_repo https://github.com/smarr/SOMns.git SOMns
+          load_git_repo https://github.com/smarr/SOMns.git SOMns dev
           pushd SOMns
           git submodule update --recursive --init
+          export JAVA_HOME=$JAVA_HOME_17_X64
+          echo "JAVA_HOME=$JAVA_HOME" >> "$GITHUB_ENV"
           ant compile
         if: matrix.id == 'somns'
 
--- a/benchmarks/SOM/CD/CollisionDetector.som
+++ b/benchmarks/SOM/CD/CollisionDetector.som
@@ -44,7 +44,7 @@ CollisionDetector = (
       newPosition := aircraft position.
       seen at: aircraft callsign put: true.
 
-      oldPosition isNil ifTrue: [
+      oldPosition ifNil: [
         "Treat newly introduced aircraft as if they were stationary"
         oldPosition := newPosition ].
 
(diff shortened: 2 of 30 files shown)


smarr:
Looking at the other languages, I think the change should be fine.

Happy to merge it, but would prefer to keep the SOM and SOMns benchmarks in sync with the Smalltalk ones.

And yeah, the splitting issue, you probably still want to fix :)

fniephaus:
Oh right, the SOM impl is separate. For some reason, I thought it’s derived from the Smalltalk code. I pushed another commit.

eregon:
There are a few more `isNil ifTrue:`, probably it would be good to change them too:
https://github.com/search?q=repo%3Asmarr%2Fare-we-fast-yet%20%2FisNil%20ifTrue%3A%2F&type=code

krono:
Note that some of these are newspeak, but the reasoning shouldn't be different.

krono:
The reason of `ifNil:` _not_ being used yet is that the original SOM, from which the benchmarks where pulled and built upon, did not support `ifNil:` back then.
Times are different now :)

fniephaus:
> There are a few more isNil ifTrue:, probably it would be good to change them too:
https://github.com/search?q=repo%3Asmarr%2Fare-we-fast-yet%20%2FisNil%20ifTrue%3A%2F&type=code

After looking at other usages, I'm not so sure anymore how far we want to go here. There are, for example, `isNil ifTrue: [] ifFalse: []` usages, which I would change to `ifNil: [] ifNotNil: []`, but that's not supported in SOM. Then, there also are usages of `notNil ifTrue: []`, which should probably be changed to `ifNotNil: []`, which is supported in SOM.

Any thoughts?

smarr:
Yeah, I pushed the SOMns commit, mostly to indicate that it's also separate code.
And, I didn't immediately merge it, because as @eregon pointed out, it would make things inconsistent. When looking at what we did for the other languages, it seemed desirable to apply this consistently to all benchmarks.

That some variations are currently missing from SOM seems only a small hurdle. Though, it's a bit of extra work. @OctaveLarose, for SOM-RS, did you implement special handling for `Boolean>>#ifNil:*`? I think I may have it for some of the SOMs already but not consistently.

OctaveLarose:
som-rs doesn't have `ifNil`, no. But it looks like it could be a speedup in a few benchmarks, and I could easily implement it. A lot of the code takes inspiration from PySOM's design, and I don't think it has any primitives for it, either.

It sounds like a good change, though! I'd not realized it was a common case.

smarr:
The update for SOM is here: https://github.com/SOM-st/SOM/pull/123

smarr (approved):

smarr:
Performance difference on the SOM implementations without specific support for `ifNil:*` variants: https://rebench.dev/Are-We-Fast-Yet/compare/2e1a50989356d8beb32c55393c8c2087d7f77707..a31e167bf1d6812a2251e10321d81abf0a3fab54

It looks a bit noisy, and there might be some extra overhead compared to the inlined `isNil ifTrue:*` version on List and Havlak.

With inlining of `ifNil:*` and variants, it seems a bit noisy and unclear, but no major change:
https://rebench.dev/Are-We-Fast-Yet/compare/2e1a50989356d8beb32c55393c8c2087d7f77707..807f8d81edef384d50638b82b80da6cb0f9e4775

PR #103, opened 2026-03-14
Fix argument count check to reflect command line help

ceresek:
Java harness help says invocation with single argument (benchmark name) is possible, using default outer and inner iteration counts, but the argument check wanted two arguments minimum.

### DIFF
--- a/benchmarks/Java/src/Harness.java
+++ b/benchmarks/Java/src/Harness.java
@@ -37,7 +37,7 @@ public final class Harness {
 
   private static void printUsage() {
     // Checkstyle: stop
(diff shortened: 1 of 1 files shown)


benchmarks/Java/src/Harness.java
smarr: This looks wrong, too.
Should be this, I think:
```suggestion
    System.out.println("Harness benchmark [num-iterations [inner-iter]]");
```

Hm, for me this suggestion shows up wrong. I am proposing to drop the brackets around [benchmark] since it's not optional.

smarr (approved):
Yeah, that's probably a bug porting from another language.
They all seem to differ in what goes into that array...

Thank you!

